PR #1, opened 2017-08-10
Used env vars to search for packages in PYTHONPATH

cftorres:
#3 @aaron-prindle @nkubala @abbytiz

### DIFF
--- a/differs/pipDiff.go
+++ b/differs/pipDiff.go
@@ -3,8 +3,10 @@ package differs
 import (
 	"io/ioutil"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"strconv"
+	"strings"
 
 	"github.com/GoogleCloudPlatform/container-diff/utils"
 	"github.com/golang/glog"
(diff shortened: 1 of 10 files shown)


differs/pipDiff.go
```diff
@@ -37,27 +39,50 @@ func getPythonVersion(pathToLayer string) ([]string, error) {
...
-		contents, err := ioutil.ReadDir(packagesPath)
+	for _, pythonVersion := range pythonVersions {
+		packagesPath := filepath.Join(path, "usr/local/lib", pythonVersion, "site-packages")
+		pythonPaths = append(pythonPaths, packagesPath)
+	}
+
+	for _, pythonPath := range pythonPaths {
+		contents, err := ioutil.ReadDir(pythonPath)
 		if err != nil {
-			// layer's Python folder doesn't have a site-packages folder
```
abbytiz: Is this no longer true in your implementation or did you just find the comment superfluous?

differs/pipDiff.go
```diff
@@ -37,27 +39,50 @@ func getPythonVersion(pathToLayer string) ([]string, error) {
 	return matches, nil
 }
 
-func (d PipDiffer) getPackages(path string) (map[string]map[string]utils.PackageInfo, error) {
-	packages := make(map[string]map[string]utils.PackageInfo)
+func getPythonPaths(vars []string) []string {
```
abbytiz: To clarify, is every path where python packages could be installed included in the environment variables?  Or is it only the path set on exit?

cftorres: What do you mean the path set on exit? 
But also, now that you mention it, I've been obtaining packages from the PYTHONPATH in addition to the default "site-packages" path, but perhaps this is wrong...  would no packages be installed in that default location if the user explicitly sets their PYTHONPATH?

abbytiz: My concern was that if a user changes their PYTHONPATH many different times and installs things in many different places, would we be finding all of those packages?

differs/pipDiff.go
cftorres: Oh hmm, you're right, if I'm preemptively building those "site-packages" paths without knowing for sure if they exist, I should just skip over them if they don't exist... I mostly deleted it because it referenced layers and presumably, there should be no layers once we address #11

differs/pipDiff.go
cftorres: Good point. I believe this would be the PYTHONPATH set at the time the container is exported into an image, so yes, if it's been overwritten and not merely added to while the container is running, you wouldn't have all the old ones... @nkubala thoughts on this?

PR #14, opened 2017-08-10
History integ tests

abbytiz:
These were lost in the docker dependency history shuffle.


abbytiz:
@nkubala @cftorres @aaron-prindle

tests/multi_hist_diff_expected.json
```diff
@@ -0,0 +1,73 @@
+[
```
cftorres: rerun for pip diff changes but otherwise lgtm

abbytiz: done, and fixed the pip differ and tests so that the path is only what is necessary.

aaron-prindle:
LGTM

PR #17, opened 2017-08-15
Save flag functional for cloud, ID, and tar

### LINKED ISSUE #6: Allow user to keep unpacked image file systems in iDiff
Currently, unpacked image file systems are automatically removed from the user's file system after the diff has been processed, but should a user wish to save those directories, we should allow them to specify that

abbytiz:
Helpful flag for debugging and looking into image file systems when diff is surprising or confusing.

@nkubala @cftorres @aaron-prindle 
fixes #6

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -128,6 +128,9 @@ var RootCmd = &cobra.Command{
 				if errMsg != "" {
 					glog.Error(errMsg)
 				}
+			} else {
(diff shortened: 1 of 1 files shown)


cftorres:
lgtm

abbytiz:
I added a log statement which should make it more obvious where to find the images if it isn't already.

PR #27, opened 2017-08-18
Changing 1.0 to 2.0 in all binary links

abbytiz:
@cftorres @aaron-prindle @nkubala


aaron-prindle:
LGTM.  Have fun in Yosemite, we'll miss you!

PR #32, opened 2017-08-18
Fix download links

duggelz:


duggelz:
Also, it seems like it would be nicer to copy the executable in to `$HOME/bin` instead of `/usr/local/bin`, and get rid of the need for `sudo`.

PR #44, opened 2017-08-25
Add version command

### LINKED ISSUE #25: Add "version" command

r2d4:
Fixes #25 

Depends on #42

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@ VERSION_MAJOR ?= 0
 VERSION_MINOR ?= 2
 VERSION_BUILD ?= 0
 
+VERSION ?= v$(VERSION_MAJOR).$(VERSION_MINOR).$(VERSION_BUILD)
+
 GOOS ?= $(shell go env GOOS)
 GOARCH = amd64
 BUILD_DIR ?= ./out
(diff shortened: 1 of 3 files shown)


aaron-prindle:
Travis failed for last commit:
```
Error: unknown command "gcr.io/gcp-runtimes/diff-base" for "To"
Run 'To --help' for usage.
exit status 1
```

r2d4:
Ah, basically the issue is that adding any subcommands makes the CLI ambiguous since the image names are passed in as unnamed args.

Might be best to remove root command usage and offer a 

`container-diff analyze ...`
`container-diff diff ...`
`container-diff version`

I could also just add it as a flag to the root command, although `-v` is already taken by glog as verbosity.

WDYT @aaron-prindle @nkubala?

aaron-prindle:
I think either is fine:
`container-diff --version`
Or refactoring so that all of the commands are subcommands.  Going forward the subcommand method makes the most sense.  Now it is going to be even more to type though lol:
`container-diff diff gcr.io/test/test:v1 gcr.io/test/test:v2`

Issue #40, opened 2017-08-24
Automatically detect whether user has local docker client installed


cftorres:
original comment in https://github.com/GoogleCloudPlatform/runtimes-common/pull/243

PR #52, opened 2017-08-28
Always use docker client

r2d4:
This removes the --eng flag and never shells out to docker.  Instead, it always uses the vendored docker client, negotiating the correct API version with the daemon.  It also updates the docker/docker dependency.


r2d4:
I'm not sure what the issue was in #40, but I think that issue shouldn't be relevant after this.
Fixes #40 

This could also probably help with https://github.com/GoogleCloudPlatform/container-diff/issues/33, since previously, the http transport for the API client wasn't being closed.

PR #60, opened 2017-08-30
split container-diff into subcommands: analyze and diff

### LINKED ISSUE #56: container-diff commands should be split into subcommands
container-diff should use subcommands for functionality instead of everything deriving from the root cmd.  Example CLI:
container-diff analyze ...
container-diff diff ...
container-diff version ... (https://github.com/GoogleCloudPlatform/container-diff/pull/44)

Ref: https://github.com/GoogleCloudPlatform/container-diff/pull/44#issuecomment-325476333

aaron-prindle:
fixes #56

### DIFF
--- a/cmd/analyze.go
+++ b/cmd/analyze.go
@@ -1,0 +1,90 @@
+package cmd
+
+import (
+	"errors"
+	"os"
+
+	"github.com/GoogleCloudPlatform/container-diff/differs"
+	"github.com/GoogleCloudPlatform/container-diff/utils"
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
+)
+
+var analyzeCmd = &cobra.Command{
+	Use:   "analyze",
+	Short: "Analyzes an image: [image]",
+	Long:  `Analyzes an image using the specifed analyzers as indicated via flags (see documentation for available ones).`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if validArgs, err := validateArgs(args, checkAnalyzeArgNum, checkArgType); !validArgs {
+			glog.Error(err.Error())
+			os.Exit(1)
+		}
+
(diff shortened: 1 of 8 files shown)


cmd/analyze.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/GoogleCloudPlatform/container-diff/utils"
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
+)
+
+var analyzeCmd = &cobra.Command{
+	Use:   "analyze",
+	Short: "Analyzes an image: [image]",
+	Long:  `Analyzes an image using the specifed analyzers as indicated via flags (see documentation for available ones).`,
+	Run: func(cmd *cobra.Command, args []string) {
```
r2d4: I'm assuming this is just a copy-paste from the analyze and diff functions?  So no need to do a full review of the code?

cmd/analyze.go
```diff
@@ -0,0 +1,98 @@
...
+		cleanupImage(image)
+		return errors.New("Could not perform image analysis")
+	}
+
+	return nil
+}
+
+func init() {
+	RootCmd.AddCommand(analyzeCmd)
+	analyzeCmd.Flags().BoolVarP(&json, "json", "j", false, "JSON Output defines if the diff should be returned in a human readable format (false) or a JSON (true).")
```
r2d4: Since analyze and diff share all the same flags, why not add a helper function here that takes a `*cobra.Command` and adds the same flag list to both of them?

aaron-prindle:
added flag sharing fxn

aaron-prindle:
This PR is to add the subcommands and change the tests, no real logic was changed.  The shared logic is in root.go currently and the diff and analyze fxns are in their respective cmd.  Will move these out to a /pkg dir and should also probably refactor utils.

PR #68, opened 2017-08-31
Changed CLI to use --types flag, a comma separated list of desired an…

aaron-prindle:


aaron-prindle:
tests and README updated.  updating `container-diff help` or other messaging to include available analyzers can be done in another PR

dlorenc:
Looks like we accidentally added a "main" file to the repo that's now conflicting. You'll have to delete.

aaron-prindle:
deleted 'main'

cmd/diff_test.go
```diff
@@ -13,7 +13,8 @@ var diffArgNumTests = []testpair{
 
 func TestDiffArgNum(t *testing.T) {
 	for _, test := range diffArgNumTests {
-		valid, err := checkDiffArgNum(test.input)
+		err := checkDiffArgNum(test.input)
```
dlorenc: This is a bit weird of a construct. Do we need err and valid? Could you do something like:

valid := (checkDiffArgNum == nil) ?

aaron-prindle: good point, changed

aaron-prindle:
the tests output the err on unexpected errors so storing it separately is needed.  changing it back

aaron-prindle:
i can just use (err == nil) in place of valid

PR #73, opened 2017-09-06
Remove fake sourceCtx

r2d4:
The containers/image library has been updated to not throw a fatal
error when it can't read /etc/docker/certs.d


aaron-prindle:
nice, LGTM

PR #75, opened 2017-09-06
Run tests in parallel, dont use working dir

r2d4:
This changes the CloudPrepper to not store intermediate artifacts in
the current directory, allowing the tests to be ran in parallel.

### DIFF
--- a/cmd/analyze.go
+++ b/cmd/analyze.go
@@ -94,8 +94,7 @@ func analyzeImage(imageArg string, analyzerArgs []string) error {
 	outputResults(analyses)
 
 	if save {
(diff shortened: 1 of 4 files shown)


aaron-prindle:
There is a '--save' flag for keeping the root file system which was previously in the current dir.  Perhaps we should change that to save-dir and make this the default for the ImagePrepper?

r2d4:
I updated the PR to print the right directory with the `--save` flag.  As for changing the name or the behavior, i think thats out of scope for this PR.  As a first step, the message should probably be printed to stdout, instead of glog info'd.

aaron-prindle:
LGTM

PR #82, opened 2017-09-12
Add release rule, exe for windows

r2d4:
I tested the release rule with my personal bucket
`RELEASE_BUCKET=somethingyouown make release`

This also automatically generates the container-diff-windows-amd64.exe binary now (it didn't have the .exe extension before)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,9 @@ BUILD_DIR ?= ./out
 ORG := github.com/GoogleCloudPlatform
 PROJECT := container-diff
 REPOPATH ?= $(ORG)/$(PROJECT)
+RELEASE_BUCKET ?= $(PROJECT)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -25,8 +25,9 @@ BUILD_DIR ?= ./out
 ORG := github.com/GoogleCloudPlatform
 PROJECT := container-diff
 REPOPATH ?= $(ORG)/$(PROJECT)
+RELEASE_BUCKET ?= $(PROJECT)
```
aaron-prindle: nit: can you default this to the release bucket we are using currently?

r2d4: PROJECT=container-diff

aaron-prindle: that makes sense but I don't have access to a container-diff project.  Just wanted to make it clear where the releases are going

PR #85, opened 2017-09-12
fix validation error for empty analyzer

aaron-prindle:


cmd/root.go
```diff
@@ -121,10 +121,13 @@ func checkArgType(args []string) error {
 }
 
 func checkIfValidAnalyzer(flagtypes string) error {
+	if len(flagtypes) == 0 {
```
r2d4: flagTypes == "" ?

aaron-prindle: done

PR #88, opened 2017-09-12
Add rule to calculate checksums of binaries

r2d4:
quick explanation of the rule
```makefile
%.sha256: %
	shasum -a 256 $< &> $@
```

`%.sha256: %` matches anything with the suffix sha256 and has a dependency on building that target

`shasum -a 256 $< &> $@` 

$< is the first dependency of the rule, in this case, the actual file we're taking the checksum of
&> redirects standard out
$@ is the name of the rule, which is the target file with the suffix .sha256

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ PROJECT := container-diff
 REPOPATH ?= $(ORG)/$(PROJECT)
 RELEASE_BUCKET ?= $(PROJECT)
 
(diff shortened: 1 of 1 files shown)


aaron-prindle:
nit: You could put a comment in the makefile regarding the make variables used if you think it'll help someone trying to modify the rule

PR #90, opened 2017-09-19
Add bazel builds for container-diff.

dlorenc:
One manual edit in vendor/gorilla/mux/BUILD.bazel to workaround a gazelle bug.


dlorenc:
cc @r2d4

BUILD
```diff
@@ -0,0 +1,25 @@
+load("@io_bazel_rules_go//go:def.bzl", "gazelle", "go_binary", "go_library", "go_prefix")
+
+gazelle(
+    name = "gazelle",
+    external = "vendored",
```
r2d4: add in 

```bazel
    build_tags = [
        "container_image_ostree_stub",
        "containers_image_openpgp",
    ],
```

dlorenc: > Should test.sh run bazel run //:gazelle? That would force any contributors to have the bazel toolchain installed.

I'm fine with that. Ultimately our downoading/installing of stuff in the minikube makefile causes extra confusion. We could try to move it into a container - but that gets tricky too...

dlorenc:
Ready for a look. Travis is passing.

Issue #86, opened 2017-09-12
Add integration test for empty/default --types


aaron-prindle:
Currently there is no integration test that checks that no --types passed results in all differs being used.  This allowed https://github.com/GoogleCloudPlatform/container-diff/pull/85 to get through.

PR #92, opened 2017-09-19
Perform apt diffing by default.

dlorenc:
Also move diff type checking up.

Before we would download all the images before failing with an invalid diff type.


dlorenc:
Ref #86

PR #104, opened 2017-09-27
Create a bazel rule for container diff

priyawadhwa:
Creates a bazel rule to diff an image built by bazel (as a tar) against an image in production with container-diff.

### DIFF
--- a/docker_diff.bzl
+++ b/docker_diff.bzl
@@ -1,0 +1,71 @@
+# Copyright 2017 Google, Inc. All rights reserved.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


docker_diff.bzl
```diff
@@ -0,0 +1,69 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+def _impl(ctx):
+  """Implementation of docker_diff"""
+
+  image_location = ctx.executable.image.short_path
+  cd_binary_loction = ctx.executable._container_diff_tool.short_path
+  
+  content = "%s diff %s %s" % (cd_binary_loction, image_location, ctx.attr.diff_base)
```
dlorenc: Could you make this a multi-line string, with a #!/bin/bash header and a set-e line?

docker_diff.bzl
```diff
@@ -0,0 +1,69 @@
...
+#     diff_types: Types flag to pass to container diff 
+#        ex: ["pip", "file"]
+
+docker_diff = rule(
+    attrs = {
+        "image": attr.label(
+            allow_files = [".tar"],
+            single_file = True,
+            mandatory = True,
+            executable = True,
```
dlorenc: Do we need executable since we're requiring the .tar?

PR #110, opened 2017-10-05
Use "latest" versions

r2d4:
* Upload to the "latest" bucket
* Point to that in the README
* Fix $(PLATFORM)-$(ARCHITECTURE) in urls


r2d4:
fixes #110

PR #116, opened 2017-10-12
Switch to logrus.

dlorenc:
This moves user output to fmt.Printf(), and logging output to logrus. It sets a default level of WARN, which can be overridden with a --verbosity flag.


dlorenc:
Whoa, hit a bunch of merge issues. Will close for now.

PR #117, opened 2017-10-12
Switch to logrus.

dlorenc:
This moves output intended for users (status style things) to fmt.Print* (stdout), and moves logging stuff to logrus (stderr). The log levels can be controlled with a --verbosity flag.


dlorenc:
OK, RFAL with tests passing.

PR #123, opened 2017-10-19
Cleanup flag parsing.

dlorenc:
This also changes the types flag from a single comma separated list to
a repeated flag that fills an array. This changes the name from types
to type, which is a minor breaking change.

Let me know what you think.

cc @aaron-prindle @nkubala


r2d4:
`container-diff diff --diff-types ...` ?  As a separate PR it might be worthwhile to rethink the commands and flags before they grow anymore...

dlorenc:
OK, tests passing now.

PR #149, opened 2017-11-27
add rpm support

vrothberg:
Extend containerdiff to support images that are shipped with rpm packages, such as SUSE or RedHat based ones.  In contrast to apt, the rpm package data is not stored in a single text file, but in Berkeley
databases, which complicates analyses.  Hence, run those images in containers to query the package data directly via `rpm -qa` to make analyses portable at the cost of requiring a docker daemon to be
installed and running.

Analyzing tar archives requires some additional work in order to run them in a container.  As multilayer tar archives generated by `docker save` must be loaded, we have to update the manifest.json file of the archive and empty all RepoTags fields.  Otherwise, docker would rename old images, which is undesired behaviour from a user's perspective.

Fixes: https://github.com/GoogleCloudPlatform/container-diff/issues/146


vrothberg:
I signed it!

differs/rpm_diff.go
```diff
@@ -0,0 +1,312 @@
...
+
+	return packages, nil
+}
+
+// loadArchiveToDaemon loads the image specified by archive to the docker
+// Daemon.
+func loadArchiveToDaemon(archive string, client *godocker.Client) (string, error) {
+	// read and write tar archive to input stream/buffer
+	inBuf := new(bytes.Buffer)
+	if content, err := ioutil.ReadFile(archive); err != nil {
```
dlorenc: nit: I think you can avoid creating the inBuf buffer and reading the entire archive into memory by passing an open File to LoadImageOptions: https://golang.org/pkg/os/#Open

differs/rpm_diff.go
```diff
@@ -0,0 +1,312 @@
...
+		InputStream:  inBuf,
+		OutputStream: outBuf,
+		Context:      context.Background(),
+	}
+	if err := client.LoadImage(loadOpts); err != nil {
+		return "", err
+	}
+
+	// output has the form "Loaded image: repository:tag"
+	output := outBuf.String()
```
dlorenc: Cool - I was wondering how you were going to handle this. It's a bit said Docker doesn't provide a way to control the image name of an image you're about to load :(

differs/rpm_diff.go
```diff
@@ -0,0 +1,312 @@
...
+	f, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
+
+	var manifestItems []dockertar.ManifestItem
+	if err := json.Unmarshal(f, &manifestItems); err != nil {
+		return nil, fmt.Errorf("error unmarshalling manifest: %s", err)
+	}
+	for i := range manifestItems {
```
dlorenc: Should we error here instead of the archive contained multiple images? It's sort of a corner-case, but the behavior will be a bit weird.

differs/rpm_diff.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+		defer os.Remove(imageName)
+
+		imageName, err = loadArchiveToDaemon(imageName, client)
+		if err != nil {
+			return packages, fmt.Errorf("Error loading archive: %s", err)
+		}
+		logrus.Debugf("Loaded archive as image %s", imageName)
+		// find out if we have to remove the image
+		inspImg, err := client.InspectImage(imageName)
```
dlorenc: Why would there be a name conflict if we remove the tags? Shouldn't docker always give us a unique ID?

differs/rpm_diff.go
```diff
@@ -0,0 +1,312 @@
...
+
+		// update the manifest
+		switch t := hdr.Typeflag; t {
+		case tar.TypeReg:
+			if hdr.Name == "manifest.json" {
+				newManifest, err := removeRepoTagsFromManifest(tr)
+				if err != nil {
+					return "", err
+				}
+				manifestHdr := hdr
```
nkubala: nit: don't need this variable

differs/rpm_diff.go
```diff
@@ -0,0 +1,312 @@
...
+
+	// if image is a tar archive, we need to load it
+	imageName := image.Source
+	if pkgutil.IsTar(imageName) {
+		// NOTE: all work below is only required for multilayer tar
+		// archives generated by `docker save`.  Flat file-system
+		// archives generated by `docker create` can just be imported
+		// without potential name conflicts.  However, those are not
+		// supported yet.
+		imageName, err = generateNewArchive(image.Source)
```
nkubala: nit: use `generateNewArchive(imageName)` since you already created the variable here

nkubala:
Thanks, this is great! As far as tests go, we have a set of images that we built and pushed to our team's GCR repository that we run different analyzers on in CI: you can see those [here](https://github.com/GoogleCloudPlatform/container-diff/blob/master/tests/integration_test.go#L38). We have the output from container-diff runs on each image saved as JSON in that same test directory, and then compare that to the output from our test runs to make sure nothing has changed.

Would you mind creating Dockerfiles for two images that have differing versions of a few RPM packages installed, then checking those in along with JSON files containing expected output of running an RPM analysis and an RPM diff on them? Once you do, I'll build and push the images to our team's GCR repo and we should be good to go.

differs/rpm_diff.go
vrothberg: There are two related issues.

The first one is to avoid name conflicts, so we have to `nil` the `RepoTags` field. This way we don't overwrite or rename any tags.

The second issue is that Docker won't give it a unique ID but a *hash*. If the hash of the loaded tar archive (layers) corresponds to the one of an already present image, we would remove it.

I just realized that checking for any tags via inspect isn't complete either as an already present image might have no tags. I just tested it and can confirm that the previously loaded image will falsely be removed. We really hit some limitations of the API here. I will see if we can avoid that using any other metadata, but even with timestamps we could run into race conditions.

Notice that we don't have this problem with a flat rootfs tar from `docker export`. The API and its semantics are much saner there.

differs/rpm_diff.go
vrothberg: Yes, we should do that. Thanks!

differs/rpm_diff.go
vrothberg: I will check that, thanks!

differs/rpm_diff.go
vrothberg: I never noticed the limitations of `docker load` until working on this PR. I understand the technical reasoning behind to reuse layers etc., but it effectively makes automation really hard - just as we're experiencing in this PR :(

differs/rpm_diff.go
vrothberg: Thanks, I will fix that.

differs/rpm_diff.go
vrothberg: Yes, this should be removed, thanks!

vrothberg:
@nkubala @dlorenc Thanks for the great reviews! I will address your comments and see how to deal with the image-removal issue. I will also prepare the tests and provide the Dockerfiles for @nkubala.

differs/rpm_diff.go
vrothberg: I put some more thoughts into the upper issue and experimented a lot, but I don't seem to converge to a clear preference. Potential solutions are:

(1) Don't remove the images of loaded archives at all and leave the janitorial work to the user.

(2) *Update* the `RepoTags` instead and give (try to give) *unique* tags in this namespace. As multiple instances of containerdiff may run in parallel, this requires using a filesystem lock or a reliable entropy library.

(3) Refuse to analyze tar archives in the RPMAnalyzer, and ask the user to load them as images before.

@nkubala @dlorenc What do you think?

vrothberg:
@nkubala @dlorenc I just pushed another commit adding the JSON test files. You can find the images and the Dockerfiles on Docker Hub( https://hub.docker.com/r/valentinrothberg/containerdiff/). Note that the 2nd commit is just there to ease review so you can spot changes more easily; we should squash it once we're done.

Regarding the `docker load` name conflicts. I opened a feature request upstream. https://github.com/GoogleCloudPlatform/container-diff/pull/149#discussion_r153433970 describes some potential ways to address the problem, and for now, I decided to always tag a tar archive's image as `containerdiff:${tar archive}`. In case we go this path, the path of the archive should be normalized to apply to docker's tag convention. What do you think?

differs/rpm_diff.go
nkubala: Good points here, this is definitely an interesting question and I'm not sure there's a perfect solution.

The most "destructive" scenario I see here is loading an archive made from exporting an old version of an image (call it `foo:latest`, with hash `abcdef`, exported to `foo.tar`), which has since been rebuilt to a new hash (`deadbeef`) and updated on the same tag in the daemon. Loading `foo.tar` into the daemon will cause the image with hash `deadbeef` to be retagged with an empty string, and image `abcdef` to be tagged as `foo:latest`, which could cause some confusion after using the tool.

Unfortunately this also means that checking for tags on the newly-loaded image won't help us know if the image needs to be left alone, because in the above scenario we just overwrote an existing image (with a tag) in the daemon, which is now an unnamed orphan image.

However, it looks to me like when doing the load, we can parse the output (like you're already doing to get the name of the newly-loaded image) and if we see something like 

`The image test:latest already exists, renaming the old one with ID sha256:<hash> to empty string`

save that hash, and then when we're done go back and retag that orphaned image with the tag it originally had, and finally remove the image we loaded in. Or at the very least, log it to the user so they know what we did. WDYT @vrothberg @dlorenc?

differs/rpm_diff.go
```diff
@@ -132,7 +124,7 @@ func rpmDataFromContainer(image pkgutil.Image) (map[string]util.PackageInfo, err
 	if err != nil {
 		return packages, err
 	}
-	logrus.Debugf("Created container %s", container.ID)
+	logrus.Info("Created container %s", container.ID)
```
nkubala: I think you want `Infof` here

differs/rpm_diff.go
vrothberg: @nkubala Thanks for getting into this!

I considered the renaming option as well, but I dropped it as retagging the old image to its initial name can run into conflicts as well. Some other image might have taken its place in the meantime. I see containerdiff to be useful in automation, so I think that post user interaction should be avoided.

Among other things, that's why I believe that using the `containerdiff:` namespace for our purpose is the most convenient solution for a user. It's more or less convention that a tool uses its name as a registry namespace for images. The code at the moment isn't perfect as it doesn't avoid name conflicts for the tags, but that's something we can protect us from by using a file system lock or a very good entropy.

So what I suggest doing is to use the `containerdiff:` namespace for any image created by containerdiff and use a file-system lock for the tagging.  I haven't used such locks until now in golang but https://godoc.org/github.com/nightlyone/lockfile makes a good impression on me; it also supports Windows. So what could happen is to lock `~/.containerdiff.lock`, generate a random tag until we hit one that isn't already used, load the image with the updated manifest, release the lock. Plus additional work to clean-up the lock in case it's previous owner died etc.

nkubala: Ah, yeah that makes sense. Would be nice if Docker provided some control over the image name though...hopefully you get some resolution on https://github.com/moby/moby/issues/35625. If we want to go the route of minimal interaction then I think your proposal should be fine. Want to give it a shot?

vrothberg: @nkubala Sure! I will try out the idea with the file-system lock tomorrow, update the PR and drop a comment explaining the change.

vrothberg:
@nkubala @dlorenc I added a new commit implementing the locking around generating a random, unique tag in the `containerdiff` namespace. It also updates the tests as requested and does some minor code cleanups. I tested the locking by forcing some race conditions via infinite loops.

I've no idea though why Travis failed: `BUILD files out of date. Run 'bazel run //:gazelle' to update them.`?

nkubala:
@vrothberg this is looking really good. It would be great if you could put your infinite loop locking test into the testsuite, just so we know the locking mechanism is working correctly.

The CI issue you're seeing is because the Bazel BUILD files are out of date. To fix this, install [bazel](https://bazel.build/), and run `bazel run //:gazelle` from the top level directory, then commit the changes.

vrothberg:
@nkubala Thanks, I am happy with the changes now although I didn't expect the tar loading to turn into something that complex :-D

The lock testing isn't as trivial as we might want it to be, which deserves some explanation (I added comments to the source and tests as well). The problem is that the lockfile is recursive-only, such that it won't protect against other goroutines, but only against other processes. I asked upstream if they are interested in adding an option to make the lock non-recursive to suit our use-case (https://github.com/nightlyone/lockfile/issues/23). Currently, I don't think that's a problem for us, but as the RPMDiffer may run in parallel in the future I played it safe and added a `sync.Mutex` to RPMDiff to also protect against other (future) routines that may run in parallel.

The recursive-lock issue means that we can't use the common `testing` infrastructure in `go test` to test `Lockfile` as we needed to fork(). Hence, the new test only tests the mutex. I suggest keeping the tests as they are now. I will track the `Lockfile` issue upstream and can update the RPMDiffer at a later point to use the non-recursive lock.

vrothberg:
@nkubala @dlorenc What do you think about the current state? If you're ready to go, I can squash the commits and provide a comprehensive commit message to have a history.

vrothberg:
@nkubala Great, thanks to you an @dlorenc for the pleasant review process! I just squashed the commits. The comment from the bot came from force pushing from a wrongly rebased commit, now on master.

PR #157, opened 2017-12-04
Search for Python version also in /usr/lib/python

fridex:
Path to Python is different on deb-based distributions and rpm-based
distributions (e.g. Fedora, CentOS).

### DIFF
--- a/differs/pip_diff.go
+++ b/differs/pip_diff.go
@@ -18,7 +18,6 @@ package differs
 
 import (
 	"io/ioutil"
-	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


differs/pip_diff.go
```diff
@@ -123,13 +124,28 @@ func addToMap(packages map[string]map[string]util.PackageInfo, pack string, path
 
 func getPythonVersion(pathToLayer string) ([]string, error) {
 	matches := []string{}
+
+	// Debian based distros
 	libPath := filepath.Join(pathToLayer, "usr/local/lib")
 	libContents, err := ioutil.ReadDir(libPath)
 	if err != nil {
-		return matches, err
+		logrus.Warnf("Could not find /usr/local/lib to determine Python version, trying /usr/lib: %s", err)
```
dlorenc: This might read a little bit cleaner if you move all of this into a for loop. Something like:

```go
// First gather all entries from both search paths
libPaths := []string{"usr/local/lib", "usr/lib"}
libContents := []os.FileInfo{}
for _, lp := range libPaths {
  libPath := filepath.Join(pathToLayer, lp)
  entries, err := ioutil.ReadDir(libPath)
  if err != nil {
    logrus.Warnf(...)
    continue
  }
  libContents = append(libContents, entries)
}

// Then look through the appended contents
for _, file := range libContents {
  ...
}
```

What do you think?

fridex:
> Made the swap to a for loop. Now it should find all the different python versions from all locations.

Thanks @dlorenc. This was on my TODO list, but I didn't find time to modify according to review comments. Thanks for refactoring!

PR #165, opened 2017-12-26
Switch to Dep

dlorenc:
The current build is actually broken on OSX right now, and I can't seem to fix it with godep.


dlorenc:
Added a note to contributing.md. I don't really want to duplicate the upstream dep docs, so I just linked to them.

PR #175, opened 2018-01-12
Add quiet flag to suppress stderr output

priyawadhwa:
#171


pkg/util/cmd_utils.go
```diff
@@ -0,0 +1,34 @@
+/*
+Copyright 2017 Google, Inc. All rights reserved.
```
dlorenc: A slightly better name/location for this might be "cmd/util/output/output.go" or something, so it's imported as "output".

pkg/util/cmd_utils.go
```diff
@@ -0,0 +1,34 @@
...
+package util
+
+import (
+	"fmt"
+	"os"
+)
+
+var quiet = false
+
+func PrintToStdErr(output string, vars ...interface{}) {
```
dlorenc: A helper to add the flag directly in here might be nice, something like:

`func AddFlags(cmd *cobra.Command)`

which could actually do the `cmd.Flags().BoolVar...` stuff. Then users wouldn't need to add the flag or set the constant.

PR #177, opened 2018-01-21
Add missing line break to README

FlorianLudwig:


FlorianLudwig:
I signed it!

Several times now :]

Issue #181, opened 2018-01-28
Add Labels to the ConfigObject


datwiz:
Add a Labels map field to the ConfigObject in pkg/util/image_prep_utils.go

Working to add metadata tests for labels in container-structure-test.  The container-structure-test tar driver uses container-diff to extract metadata from the image tar file.

PR #182, opened 2018-01-28
added Labels to the ConfigObject

datwiz:
Added Lables to the image ConfigObject.
Resolves Issue #181


datwiz:
Tests passing in my local build environment, but failing in TravisCI build.  Working on isolating the cause.

datwiz:
Thanks,  I think the addition is still causing a build failure and the green build status may have been a race between the green build where I removed my change and the red build when I added it back.
The issue I tracking down is whether there is a difference in json unmarshalling between the environment I'm using locally and the travis build environment.

datwiz:
Just re-ran a build from my github repo and looks like the build was successful this time.

PR #191, opened 2018-02-15
modifications to extract filesystem for kbuild

priyawadhwa:
Extract symlinks from the tarball, and also add support for a whitelist (files that would be ignored during extraction)

### DIFF
--- a/pkg/util/cloud_prepper.go
+++ b/pkg/util/cloud_prepper.go
@@ -59,7 +59,7 @@ func (p CloudPrepper) GetFileSystem() (string, error) {
 		return "", err
 	}
 
-	return path, GetFileSystemFromReference(ref, p.ImageSource, path, []string{})
+	return path, GetFileSystemFromReference(ref, p.ImageSource, path, nil)
 }
 
 func (p CloudPrepper) GetConfig() (ConfigSchema, error) {
--- a/pkg/util/cloud_prepper.go
+++ b/pkg/util/cloud_prepper.go
@@ -59,7 +59,7 @@ func (p CloudPrepper) GetFileSystem() (string, error) {
 		return "", err
 	}
 
-	return path, getFileSystemFromReference(ref, p.ImageSource, path)
+	return path, GetFileSystemFromReference(ref, p.ImageSource, path, []string{})
 }
 
 func (p CloudPrepper) GetConfig() (ConfigSchema, error) {
--- a/pkg/util/docker_utils.go
+++ b/pkg/util/docker_utils.go
@@ -115,7 +115,7 @@ func unpackDockerSave(tarPath string, target string) error {
 	}
 
 	for _, layer := range layers {
-		if err = UnTar(bytes.NewReader(layerMap[layer]), target); err != nil {
+		if err = UnTar(bytes.NewReader(layerMap[layer]), target, []string{}); err != nil {
 			return fmt.Errorf("Could not unpack layer %s: %s", layer, err)
 		}
 	}
--- a/pkg/util/docker_utils.go
+++ b/pkg/util/docker_utils.go
@@ -115,7 +115,7 @@ func unpackDockerSave(tarPath string, target string) error {
 	}
 
 	for _, layer := range layers {
-		if err = UnTar(bytes.NewReader(layerMap[layer]), target, []string{}); err != nil {
+		if err = UnTar(bytes.NewReader(layerMap[layer]), target, nil); err != nil {
 			return fmt.Errorf("Could not unpack layer %s: %s", layer, err)
 		}
 	}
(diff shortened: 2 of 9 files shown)


pkg/util/cloud_prepper.go
```diff
@@ -59,7 +59,7 @@ func (p CloudPrepper) GetFileSystem() (string, error) {
 		return "", err
 	}
 
-	return path, getFileSystemFromReference(ref, p.ImageSource, path)
+	return path, GetFileSystemFromReference(ref, p.ImageSource, path, []string{})
```
dlorenc: You should be able to put nil here.

pkg/util/docker_utils.go
```diff
@@ -115,7 +115,7 @@ func unpackDockerSave(tarPath string, target string) error {
 	}
 
 	for _, layer := range layers {
-		if err = UnTar(bytes.NewReader(layerMap[layer]), target); err != nil {
+		if err = UnTar(bytes.NewReader(layerMap[layer]), target, []string{}); err != nil {
```
dlorenc: And here.

pkg/util/tar_utils.go
```diff
@@ -65,7 +67,7 @@ func unpackTar(tr *tar.Reader, path string) error {
 				if err := os.MkdirAll(target, mode); err != nil {
 					return err
 				}
-			} else {
+				// In some cases, MkdirAll doesn't change the permissions, so run Chmod
```
dlorenc: It's usually umask related, but I suppose if the dirs already exist it wouldn't change the permissions either

pkg/util/tar_utils.go
```diff
@@ -106,21 +107,52 @@ func unpackTar(tr *tar.Reader, path string) error {
...
+	return filepath.Walk(p, func(path string, info os.FileInfo, err error) error {
+		if e := os.Chmod(path, 0777); e != nil {
+			logrus.Errorf("Error updating file permissions on %s before removing for symlink creation", path)
+			return e
+		}
+		return os.RemoveAll(path)
+	})
+}
+
+func checkWhitelist(target string, whitelist []string) bool {
```
dlorenc: I think you want something more precise than prefix matching. Something like /foo would block /foobar when it should only block /foo/bar

pkg/util/fs_utils.go
```diff
@@ -160,3 +160,22 @@ func CheckSameFile(f1name, f2name string) (bool, error) {
 	}
 	return true, nil
 }
+
+// HasFilepathPrefix checks if the given file path begins with prefix
+func HasFilepathPrefix(path, prefix string) bool {
+	path = filepath.Clean(path)
+	prefix = filepath.Clean(prefix)
+	pathArray := strings.Split(path, "/")
```
dlorenc: Can you use filepath.Split here?

pkg/util/tar_utils.go
```diff
@@ -49,13 +47,16 @@ func unpackTar(tr *tar.Reader, path string) error {
...
-			if err = os.RemoveAll(newName); err != nil {
+			if err = walkAndRemove(newName); err != nil {
 				logrus.Error(err)
 			}
 			continue
 		}
-
 		target := filepath.Join(path, header.Name)
+		// Make sure the target isn't part of the whitelist
+		if checkWhitelist(target, whitelist) {
```
dlorenc: A debug log here will come in handy later.

pkg/util/tar_utils.go
```diff
@@ -106,21 +106,52 @@ func unpackTar(tr *tar.Reader, path string) error {
...
+				logrus.Errorf("Failed to create symlink between %s and %s: %s", header.Linkname, target, err)
+			}
+		}
 	}
 	return nil
 }
 
+func walkAndRemove(p string) error {
+	return filepath.Walk(p, func(path string, info os.FileInfo, err error) error {
+		if e := os.Chmod(path, 0777); e != nil {
```
dlorenc: Why do we need to chmod before removal?

priyawadhwa: Ah sorry, I was trying something out and meant to remove that

pkg/util/fs_utils.go
priyawadhwa: I think it only [splits](https://golang.org/pkg/path/filepath/#Split) along the last separator

pkg/util/tar_utils.go
priyawadhwa: I added it in the checkWhitelist function

pkg/util/fs_utils.go
dlorenc: D'oh. Right.

PR #200, opened 2018-03-09
Add createdBy field to mutable source config history

priyawadhwa:


pkg/image/mutable_source.go
```diff
@@ -177,28 +177,29 @@ func (m *MutableSource) Env() map[string]string {
...
-	m.AppendConfigHistory(author, true)
+	m.AppendConfigHistory(author, createdBy, true)
 }
 
 func (m *MutableSource) Config() *manifest.Schema2Config {
 	return m.cfg.Schema2V1Image.Config
 }
 
-func (m *MutableSource) SetConfig(config *manifest.Schema2Config, author string, emptyLayer bool) {
+func (m *MutableSource) SetConfig(config *manifest.Schema2Config, author, createdBy string, emptyLayer bool) {
```
dlorenc: Maybe refactor this a bit to take a manifest.Schema2History struct instead of all the fields as parameters.

PR #220, opened 2018-04-10
Update populateManifestAndConfig to work with docker registry images

priyawadhwa:
kaniko wasn't able to build from images in the Docker registry because it wasn't able to find the manifest or config. Instead of getting a manifest, populateManifestAndConfig first gets a manifest list for these images, so I added support for that. I also changed how we get the config so that it would work with Docker and GCR images.

### DIFF
--- a/pkg/image/mutable_source.go
+++ b/pkg/image/mutable_source.go
@@ -20,13 +20,14 @@ import (
 	"bytes"
 	"compress/gzip"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


priyawadhwa:
So I specified the context (linux/amd64) but it looks like calling Manifest() still gives you a manifest list, so I still have to call ChooseManifestInstanceFromManifestList() to pick out the manifest with the context we want.

PTAL :)

PR #240, opened 2018-06-08
Do not overwrite the loaded tarball image

davidcassany:
This commit makes sure that if the provided argument is a local
tarfile, the loaded image into the `img` varible does not get
overwritten by trying to fetch something from a remote registry.

From what I understand the `img` variable initialized in https://github.com/GoogleContainerTools/container-diff/blob/0ffd08ab6b87fa4997c0f6a62dfb1371dea02ad9/cmd/root.go#L132 is always ignored by assuming the given argument is a remote reference.


davidcassany:
I signed it!

davidcassany:
I signed it!

Hopefully with correct settings this time

PR #247, opened 2018-07-12
Reuse cached filesystems for layers

davidcassany:
This commit applies the same criteria for reusing extracted layer
filesystems than the one used for the whole image extracted filesystem


davidcassany:
Without this patch layer analysis was not reusing any cached layer, moreover in some cases it was even failing because of trying to extract the layer in a folder where it was already extracted previously, causing issues with non writable files and symlinks.

PR #286, opened 2019-01-09
Adding Github Action to run Container Diff

vsoch:
This pull request will add a folder, "actions" that has inside a Dockerfile, entrypoint, and instructions for using container-diff within Github Actions. The simplest example is just to run container-diff as part of a workflow:

```
workflow "Run container-diff isolated" {
  on = "push"
  resolves = ["Run container-diff"]
}
action "Run container-diff" {
  uses = "GoogleContainerTools/container-diff/actions@master"
  args = ["analyze vanessa/salad --type=file --output=/github/workspace/data.json --json"]
}
```

You could also list the data.json to show it's available for the next step:

```
workflow "Run container-diff isolated" {
  on = "push"
  resolves = ["list"]
}
action "Run container-diff" {
  uses = "GoogleContainerTools/container-diff/actions@master"
  args = ["analyze vanessa/salad --type=file --output=/github/workspace/data.json --json"]
}
action "list" {
  needs = ["Run container-diff"]
  uses = "actions/bin/sh@master"
  runs = "ls"
  args = ["/github/workspace"]
}
```
Which is what I did during testing:

image

And the general expectation is that the user would then do some additional step in the workflow that uses the data. For example, I'm going to use it to create an interactive tree diagram of the container (using the --type=file output) and then push it back to github pages.

I don't see need for different kinds or levels of actions (e.g., a subfolder in a repository can have a different Dockerfile or with a different branch) so I made just one actions folder with the Dockerfile at the root, but if you see reason to add specific ones let's chat about that.

### DIFF
--- a/actions/Dockerfile
+++ b/actions/Dockerfile
@@ -1,0 +1,44 @@
+FROM golang:1.11.3-stretch
+
+# docker build -f actions/Dockerfile -t googlecontainertools/container-diff .
+
+RUN apt-get update && \
+    apt-get install -y automake \
+                       libffi-dev \ 
+                       libxml2 \
+                       libxml2-dev \
+                       libxslt-dev \
+                       libxslt1-dev \
+                       git \
+                       gcc g++ \
+                       wget \
+                       locales
+
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
+    locale-gen
+ENV LANG en_US.UTF-8  
+ENV LANGUAGE en_US:en  
+ENV LC_ALL en_US.UTF-8
+
+LABEL "com.github.actions.name"="container-diff GitHub Action"
+LABEL "com.github.actions.description"="use Container-Diff in Github Actions Workflows"
+LABEL "com.github.actions.icon"="cloud"
+LABEL "com.github.actions.color"="blue"
+
+LABEL "repository"="https://www.github.com/GoogleContainerTools/container-diff"
+LABEL "homepage"="https://www.github.com/GoogleContainerTools/container-diff"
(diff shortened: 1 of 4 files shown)


vsoch:
I'm not familiar with the setup with GoLang, but it looks like some kind of boilerplate (or ignore file?) is missing here?

```
Checking boilerplate...
Boilerplate missing in: ././actions/Dockerfile
././actions/entrypoint.sh.
make: *** [test] Error 1
/home/travis/.travis/job_stages: line 523:  3893 Terminated              travis_jigger "${!}" "${timeout}" "${cmd[@]}"
The command "travis_wait make test integration" exited with 2.
```
Let me know how to fix this up.

vsoch:
oup, found it! Disregard question :)

vsoch:
Awesome! If you have some random free time, definitely try them out. It's akin to any continuous integration, but all within Github (so if you needed to share your `GITHUB_TOKEN` with an external service, you wouldn't need to. Any repository can provide an action, in any subfolder, and the Dockerfile in that subfolder drives it! So for container-diff, I added the "actions" folder, and within someone's workflow (in a different repository) they would reference the action like:

```
workflow "Run container-diff isolated" {
  on = "push"
  resolves = ["Run container-diff"]
}
action "Run container-diff" {
  uses = "GoogleContainerTools/container-diff/actions@master"
  args = ["analyze vanessa/salad --type=file --output=/github/workspace/data.json --json"]
}
```
The "uses" key references the repository, and the subfolder "actions" on the master branch. The "args" is where they write the command they want to send to container-diff. 

In context of a workflow, this would just be one step! Likely next they would do something with the output. We aren't biased to what is done, we just want to make sure they can easily do it :)

PR #304, opened 2019-03-05
version: Move vX.Y.Z to version.go so it works with `go get`, add git info

cben:
Followup from #293.
Previously, building with `make` was required for `version` command to output "v0.15.0"; if you just `go get` or clone and `go build .` it would always tell you "v0.0.0-unset" :-(

Now the basic vX.Y.Z is always available, and when building with `make` git information is appended:
```
$ go run . version
v0.15.0 built from git (unknown)
$ out/container-diff version
v0.15.0 built from git v0.15.0-1-gcd5a837-dirty
```
(I added `version --short` mode for `make release` to be able to extract the const embedded in version.go without anything else.  IMHO that's cleaner than slicing the output of `version` but could go either way.)

### Alternatives:
https://github.com/golang/go/issues/22147, https://github.com/golang/go/issues/29814 explore several directions for letting binaries know which git they were built from.  ldflags is indeed main current way.
Most of these don't work with bare `go get`.  
(`go generate` is not an option as go get won't run it, it's intended for the author to run and commit)

Go 1.12 has new `debug.ReadBuildInfo()` api that's supposed to automagically provide build info without needing ldflags, but only if you use go modules (?) and people reported some issues using it: https://github.com/golang/go/issues/29228.  See also https://github.com/golang-migrate/migrate/pull/172 for example attempt at using that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,12 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Bump these on release
-VERSION_MAJOR ?= 0
-VERSION_MINOR ?= 15
-VERSION_BUILD ?= 0
-
-VERSION ?= v$(VERSION_MAJOR).$(VERSION_MINOR).$(VERSION_BUILD)
+# On release, remember to also bump const in version/version.go
+GIT_VERSION ?= $(shell git describe --always --tags --long --dirty)
 
 GOOS ?= $(shell go env GOOS)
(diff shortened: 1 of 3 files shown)


cben:
BTW, I don't think README says how to build from source, initially I didn't notice the Makefile and assumed `go get` / `go build`, which is what led me think `version` command is completely useless :-)

PR #307, opened 2019-04-01
README: mention archlinux-specific instructions

SantiagoTorres:
This adds a blurb mentioning instructions to install container-diff under arch
linux in the README :)


SantiagoTorres:
I signed it!

PR #310, opened 2019-05-22
Revert "Add a cloudbuild.yaml for tag releases."

termux-io:
Reverts GoogleContainerTools/container-diff#126


termux-io:
Https://blockspin.io

PR #324, opened 2020-02-02
Fix concurrent map write for hardlink

antechrestos:
This change uses `sync.Map` as suggested in #323 . It also uses a `sync.atomic.Value` container for any error that might happens while loop iteration

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 out/*
+.idea/
+*.iml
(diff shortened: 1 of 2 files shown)


antechrestos:
@nkubala any feedback on this?
thanks

antechrestos:
@nkubala no problem smil9

PR #325, opened 2020-02-11
Document that container-diff is not officially supported

tstromberg:


tstromberg:
I like your wording much better.  Thank you!

PR #349, opened 2021-02-12
Remove global variable

leonor16resende:
This PR is a small code improvement. Since the variable `hardlinks` is only used in the function `unpackTar()`, there is no need for it to have global scope. Therefore, this change will limit the variable scope to this function only.


leonor16resende:
ping @nkubala @tstromberg

nkubala (approved):
thanks @leonor16resende!

PR #359, opened 2021-06-08
Fix short help with analyzeCmd and diffCmd

blueyed:
Fixes "unknown command" errors when using it as described previously.


nkubala (approved):
@blueyed thanks!

PR #360, opened 2021-06-25
Make installation command be better

lvjp:
When I tried to install `container-diff` on my computer, I wasn't able to see the Linux install command without scrolling.

I thought that it would be more convenient to make command multiline.

And, by the way, I made a little simplification with the usage of the well-known command `install`.


nkubala (approved):
@lvjp thanks!

Issue #375, opened 2022-03-17
Wrong diff labels output 'entries have been added to ...'


gabyx:
### Expected behavior
Correct naming what is added **where**.

### Actual behavior
It reports something has been added to imageA, but its the other way round.
Output should be:

`These entries have been added to bbbbb:latest:`

The template is probably just wrong. 

### Information

- container-diff version: v0.17.0 built from git bea7ad9
- Operating system: macOS

### Steps to reproduce the behavior

1.  ```dockerfile
    FROM alpine:latest as A
    RUN touch A.txt

    FROM alpine:latest as B
    RUN touch B.txt
    ```

2. ```shell
   docker build -f Dockerfile --target b -t bbbbb:latest . && \
   docker build -f Dockerfile --target b -t aaaaa:latest . 
   ```
3. `container-diff diff daemon://aaaaa:latest daemon://bbbbb:latest --type=file`

4. Output is misleading:
   ```
    -----File-----

    These entries have been added to aaaaa:latest:
    FILE          SIZE
    /B.txt        0

    These entries have been deleted from aaaaa:latest:
    FILE          SIZE
    /A.txt        0
    ```

AaronLuo3296:
Seems this issue was resolved. Would you please make a new release? Thanks.

PR #376, opened 2022-03-17
Wrong diff output

gabyx:
Wrong diff output. See issue #375.


tequilarista (approved):

felipecrs:
A new patch release with this fix would be awesome... :)

If I can help in anything, like filling the CHANGELOG or so, please let me know.

PR #387, opened 2023-01-20
Add php_diff to list differences in php extensions

rialg:
Analyzer for PHP extensions and differences in sizes. This analyzer checks for installed dynamic libraries under the `PHP_INI_DIR` directory.


rialg:
Hi @tequilarista, is this repo actively maintained? I'd like to extend it to compare differences between php modules/packages/extensions, would it be possible?

TBG-FR:
> Hi @tequilarista, is this repo actively maintained? I'd like to extend it to compare differences between php modules/packages/extensions, would it be possible?

Looks like nobody is answering... What did you end up doing ?

PR #398, opened 2023-08-24
docs: add brew way to readme

Dentrax:
Adds `brew` installation way to README.

cc @paul-hammant


loosebazooka:
Uh sorry, this has been ignored. Is this still valid?

PR #401, opened 2023-10-03
Create GitHub action.yml

MPV:
## What

Add a proper `action.yml` file for GitHub Actions support, building on the previous work by @vsoch in:
- https://github.com/GoogleContainerTools/container-diff/pull/286

As per docs:
- https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-docker-container-actions

## Why

So it's possible to easily use this action in your workflows, like this:

```yaml
[...]
steps:
  [...]
  - uses: GoogleContainerTools/container-diff@master
```

PR #402, opened 2023-10-11
Update GitHub Action for Modern Syntax

vsoch:
This PR is based off of #401 to honor the start of the effort by @MPV. I am adding to his start the following:

- Updating the Dockerfile for a smaller base image (and one that builds!) and one that downloads the latest release (and does not need to go build)
- Adding a local test for the action to run on a pull request
- The action.yaml should be in the actions subdirectory
- Updating the action arguments to support command and args. Command is by default "analyze" and args can be whatever the user needs for the command in question. The split isn't required, but I think it's good to provide a bit of structure. Note that arguments for container actions don't need to be in the action.yml, they can be derived from the environment via `INPUT_<NAME>`
- Updating docs in the actions/README.md and adding a table with said arguments.

Please let me know what you would like to discuss / change, happy to do that!

### DIFF
--- a/.github/workflows/test-action.yaml
+++ b/.github/workflows/test-action.yaml
@@ -1,0 +1,24 @@
+name: Test container-diff Action
+
+on: 
+  pull_request: []
+
+jobs:
+  test-container-diff:
+    name: Test container-diff
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # Add more test cases here as necessary
+        args:
+          - vanessa/salad --type=file --output=./data.json --json
+    steps:
+      - name: Checkout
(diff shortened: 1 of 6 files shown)


MPV (approved):
Yay, great stuff! 🥰

vsoch:
It's very quick and ensures we control the base environment. What you pointed out was a composite action that runs directly on the host. I think my preference is to maintain what we currently have, and have another PR that follows up with a larger change if desired. Asking for more than that is a bit of scope creep imho.

MPV:
No worries, of course up to you to decide what makes sense to include and maintain in this context. 👍 As said before, the PR looks good and I guess there's just an approval needed to merge.

loosebazooka (approved):

